PR #1, opened 2016-08-17
Add interfaces from the current draft of PSR-15

mnapoli:
It will be useful to start playing with it.

I've added the interfaces as-is from the current version of the draft.


shadowhand:
👍

shadowhand:
Thanks!

PR #4, opened 2016-08-18
Rename Frame to Delegate

shadowhand:
Attempt to resolve confusion as to the purpose of the "next" caller.

### DIFF
--- a/src/DelegateInterface.php
+++ b/src/DelegateInterface.php
@@ -5,7 +5,7 @@ namespace Psr\Http\Middleware;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
-interface FrameInterface
+interface DelegateInterface
 {
     /**
      * Dispatch the next available middleware and return the response.


shadowhand:
@mnapoli @mindplay-dk is this enough to satisfy the immediate concern regarding the name of the "next middleware" dispatcher?

mindplay-dk:
I still would go with full-on `MiddlewareDelegateInterface` to be explicit - I know it's longer, but should raise less questions ("delegate of what?") and it's not like we'll be typing that interface name all day? ;-)

shadowhand:
@mindplay-dk not sure I agree the interface needs to be longer. The interface is clearly in the HTTP namespace and will only be used in a middleware.

mindplay-dk:
> The interface is clearly in the HTTP namespace and will only be used in a middleware.

Okay, good point.

mnapoli:
Should the same change be reflected in PSR-15 and its meta document?

shadowhand:
@mnapoli yes it should.

mnapoli:
From your answer I didn't know if you already worked on it or planned to work on it, so … https://github.com/php-fig/fig-standards/pull/808

shadowhand:
@mnapoli was planning on submitting a PR in the next 48 hours. Glad someone else took initiative. 😄

Issue #3, opened 2016-08-18
Allow Stack to implement Middleware


shadowhand:
An interesting suggestion made by @mnapoli was that a middleware Stack could be used as a Middleware to allow deep chaining. In order to achieve this, Stack and Middleware would need to have compatible signatures. Currently both define a `process()` method with an incompatible signature.

Is this something that should be allowed?

If it is allowed, what would be appropriate names for each method?

shadowhand:
Personally I am in favor of allowing Stack to be a Middleware. For naming:

```
Middleware::process($request, $next);
Stack::dispatch($request);
```

Thoughts @http-interop/http-middleware-contributors ?

mindplay-dk:
I already implement dispatcher as middleware - I believe it's important to achieve better modularity. (you can [see here](https://github.com/mindplay-dk/middleman/blob/lambda-style/src/Dispatcher.php#L67) how I did it in my dispatcher.)

I've noticed something interesting about what's currently referred to as the "stack" interface - it's actually identical to a controller interface:

``` php
interface Controller {
    public function dispatch(RequestInterface $request): ResponseInterface;
}
```

"Takes a Request and produces a Response" is probably the broadest, most general definition you can have of a controller interface?

This was one of the reasons I changed my mind about lambda vs double-pass - it fits much better with controllers. (whereas with double-pass, would-be controllers had to be "response builders" which were given the existing incoming Response and had to populate it, which seemed wrong.)

Either way, what I'm thinking is that this pattern is so common and so generally useful, we should not refer to it as a "stack interface" - a stack happens to be a controller, but so do many other types.

shadowhand:
@mindplay-dk I don't want to mix middleware with MVC. In the case of client middleware, calling it a controller doesn't make sense. Those of us that use ADR will find it confusing.

The "stack" name also has foundation in the StackPHP project.

mindplay-dk:
> I don't want to mix middleware with MVC

Understood and agreed.

> In the case of client middleware, calling it a controller doesn't make sense

Okay, but calling it a "stack" doesn't make sense either.

The ability to dispatch a request and return a response does not make something a stack. The methods previously present were about manipulating a stack - the remaining interface method signature has nothing directly to do with being a stack.

So how about a name that describes only the remaining responsibility, e.g.`Dispatchable`?

My concern remains though, that this interface signature is so general, it's not even middleware-specific in the first place - having multiple identical interfaces might create interoperability problems.

[...]

PR #5, opened 2016-08-18
Allow Stack to implement Middleware

shadowhand:
Being able to use stacks as middleware allows for complex compositions.

Refs #3

### DIFF
--- a/src/StackInterface.php
+++ b/src/StackInterface.php
@@ -8,33 +8,7 @@ use Psr\Http\Message\ResponseInterface;
 interface StackInterface
 {
     /**
-     * Return an instance with the specified middleware added to the stack.
-     *
-     * This method MUST be implemented in such a way as to retain the
-     * immutability of the stack, and MUST return an instance that contains
-     * the specified middleware.
-     *
-     * @param MiddlewareInterface $middleware
-     *
-     * @return self
-     */
-    public function withMiddleware(MiddlewareInterface $middleware);
-
-    /**
-     * Return an instance without the specified middleware.
-     *
-     * This method MUST be implemented in such a way as to retain the
-     * immutability of the stack, and MUST return an instance that does not
-     * contain the specified middleware.
-     *
-     * @param MiddlewareInterface $middleware
-     *
-     * @return self
-     */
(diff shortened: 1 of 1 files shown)


src/StackInterface.php
```diff
@@ -43,5 +17,5 @@ public function withoutMiddleware(MiddlewareInterface $middleware);
      *
      * @return ResponseInterface
      */
-    public function process(RequestInterface $request);
+    public function dispatch(RequestInterface $request);
```
harikt, line 20: What is the reason behind the change in method name ?

mnapoli, line 20: @harikt See the issue.

harikt, line 20: @mnapoli yes I did go through the groups discussion and issue. Probably I didn't quite understand whether change in method name have any benefit. I will go though it once again to understand the reasoning.

mnapoli, line 20: Here is for example an implementation of a middleware pipe _as a middleware_: https://github.com/stratifyphp/http/blob/master/src/Middleware/Pipe.php#L36

With the current interface a pipe/stack cannot be a middleware at the same time because the methods have the same name.

I don't know if a lot of projects do that, but this is really great to compose middlewares and create complex architectures, here is a very basic example: https://gist.github.com/mnapoli/4b082c1a9812904cddc1a5719d01e889 (as you can see the router is also a middleware)

mindplay-dk, line 20: > With the current interface a pipe/stack cannot be a middleware at the same time because the methods have the same name.

That's problematic, I agree.

shadowhand:
@mindplay-dk if you're in favor, can you approve the PR please?

src/StackInterface.php
markstory: Pipelines as middleware is something we're exploring [for CakePHP](https://github.com/cakephp/cakephp/issues/9089) as well. It has some nice benefits in that pipelines and individual middleware are easily composed together.

mindplay-dk:
@shadowhand with `StackInterface` removed by #12, this PR is no longer relevant, correct?

shadowhand:
@mindplay-dk yes.

PR #8, opened 2016-08-18
Change root namespace to Interop

shadowhand:
This package does not (at this time) represent an accepted PSR.

### DIFF
--- a/src/DelegateInterface.php
+++ b/src/DelegateInterface.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Psr\Http\Middleware;
+namespace Interop\Http\Middleware;
 
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
--- a/src/MiddlewareInterface.php
+++ b/src/MiddlewareInterface.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Psr\Http\Middleware;
+namespace Interop\Http\Middleware;
 
 use Psr\Http\Message\RequestInterface;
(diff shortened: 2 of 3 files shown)


harikt:
As @weierophinney already approved for http-factory why not merge this also?

mnapoli:
:+1: I thought that that package was a strict representation of the PSR

weierophinney:
I'm not part of either team as a sponsor or contributor, @harikt; I was
only expressing approval/agreement.

shadowhand:
@http-interop/http-middleware-contributors one of us needs to sign off on this.

shadowhand:
@mindplay-dk please sign off on this so that I can merge and tag a release.

shadowhand:
@http-interop/http-middleware-contributors someone please approve this so that we can put it on packagist.

mindplay-dk:
For the record, I don't agree with this change.

But I seem to be the minority, so: approved so we can move ahead.

weierophinney:
@mindplay-dk Care to elaborate? I'm curious why you don't agree...

mindplay-dk:
@weierophinney just that, it seems like we're assuming failure. I'd rather assume success, and not have to make a breaking change if the PSR is approved, by then changing the namespace back again. If the PSR is rejected, that _is_ a breaking change, and a namespace change would then be a natural change.

If this was no longer a PSR, the namespace should of course change. At the moment, it _is_ a PSR, albeit unstable, so the PSR namespace seems appropriate.

Anyhow, bigger fish to fry, hence approved in the name of progress :-)

PR #12, opened 2016-08-23
Remove empty middleware interface

### LINKED ISSUE #11: Middleware vs "server" middleware
I'd like to discuss options for resolving the weird situation with the empty interface.

An empty interface does not provides type-safety, since it does not define any behavior, capabilities or responsibilities. The empty interface is, in my view, a symptom that we're doing something wrong.

For example, type-hinting as `MiddlewareInterface` in a dispatcher does _not_ provide auto-completion or inspections in an IDE; you still need `if`/`instanceof`-statements in dispatchers to determine the actual type of middleware - which is no different from the situation today - so the type-hint itself doesn't provide any type-safety at all.

In client/server-middleware, the proposed middleware-signature also does not provide type-safety for neither client nor server-requests, nor even for the dispatch-method, at all.

Here are some ideas.

**OPTION 1: Single Interface**

The obvious option is to accept the lack of type-hintability of server-specific middleware.

This isn't the same as ignoring it, and it's not necessary the wrong approach.

I've noticed, currently, my own server-middleware follows this pattern:

``` php
    public function __invoke(RequestInterface $request, callable $next = null)
    {
        if ($request instanceof ServerRequestInterface) {
            // ...
        }

        return $next($request);
    }
```

This doesn't seem wrong? I mean, in a world with a single middleware interface, for middleware that works both on the client and server, this seems the correct thing to do.

So I think one option, is to simply have only a single interface - consistent with what we have in the de-facto standard today.

**OPTION 2: Separate Interfaces**

Another option is to not have a third "general middleware" interface, at all.

The methods would need dedicated names:

``` php
interface ClientMiddlewareInterface {
    public function processClientRequest(RequestInterface $request, DelegateInterface $next);
}

interface ServerMiddlewareInterface {
    public function processServerRequest(ServerRequestInterface $request, DelegateInterface $next);
}
```

Note that the interface method-names are different. (using long method-names here to clarify for discussion, but one could rationalize shorter names like `process` for client-middleware and `dispatch` for server-middleware, if you can accept the idea that client-middleware doesn't actually dispatch requests per se - as opposed to server-middleware which is actually dispatches requests.)

Client and server middleware, respectively, would implement the relevant interface, and client/server-middleware would implement both interfaces. This approach eliminates the use of if/else-statements in middleware that supports both client and server.

In dispatchers, we would need to type-hint with php-doc only, but as argued above, the empty middleware-interface requires exactly the same run-time type-checks in dispatchers now - and it's an isolated problem anyhow, by which I mean, there are much fewer dispatcher than middleware implementations; the two if-statements required to type-check incoming middleware are not going to be a big problem for dispatchers - any type-related issue is a much bigger burden to middleware authors.

Footnote: if we have support for [union types](https://wiki.php.net/rfc/union_types) at some point, this problem will melt away.

**OPTION 3: Explicit Third Interface**

One last option, perhaps the most technically correct, is to have completely separate interfaces and method-names for all three cases, requiring middleware implementers to explicitly indicate whether a middleware component is client, server, or client/server-compatible; and in client/server-middleware, requiring them to explicitly implement dispatch for all _three_ (not two) cases.

So the interfaces (very wordy here to further understanding) would look like:

``` php
interface ClientMiddlewareInterface {
    public function processClientRequest(RequestInterface $request, DelegateInterface $next);
}

interface ServerMiddlewareInterface {
    public function processServerRequest(ServerRequestInterface $request, DelegateInterface $next);
}

interface ClientServerMiddlewareInterface
    extends ClientMiddlewareInterface, ServerMiddlewareInterface
{}
```

Note the key difference from what we have now, is that the client/server-middleware interface actually has two methods, and requires the middleware authors to implement both methods explicitly - eliminating the `if`/`instanceof` run-time type-check in middleware.

I don't want to point at one of these options as my favorite - I do have a favorite, but I'd like to check the pulse on @http-interop/http-middleware-contributors before coloring your opinions.

Thoughts?

shadowhand:
Empty middleware interface does nothing only existed as a convenience
for stack/dispatch. By removing the stack interface, we no longer need
the empty interface.

This also removes the need to have a "client" middleware interface.
Since all "client" middleware can be used in server context, there is
no need to imply that it is "client-only".

Refs #3
Fixes #11


mindplay-dk:
Is this ready to merge?

I'd like to try it out on middleman and see how that pans out.

mnapoli:
:+1: this is so much simpler \o/

PR #25, opened 2016-11-06
Update README

shadowhand:
Rework document based on current interfaces. Add more detail regarding
design decisions and clarify intent to focus on server middleware.

PR #26, opened 2016-11-06
Import PSR meta as README

shadowhand:
Importing the meta document from PSR-15 for review and editing.

Refs #25


shadowhand:
This is the current meta document from [php-fig/fig-standards PSR-15-meta](https://github.com/php-fig/fig-standards/blob/154fe192d102954434e2d9bac0bcaa3a7b5b4a23/proposed/http-middleware/middleware-meta.md).

README.md
```diff
@@ -1,3 +1,226 @@
-# HTTP Middleware
+HTTP Middleware
```
weierophinney: Was going to suggest changing the above to "HTTP Server Middleware", and then realized #25 is a PR against this branch, doing that...

README.md
```diff
@@ -1,3 +1,226 @@
...
+```
+fn(request, response, next): response
+```
+
+Based on the middleware implementations already used by frameworks that have
+adopted this signature, the following commonalities are observed:
+
+* The middleware is defined as a [callable][php-callable] using `__invoke`.
+* The middleware is passed 3 arguments during invocation:
+  1. A `RequestInterface` implementation.
```
weierophinney: If we're talking server-side middleware, this would be a `ServerRequestInterface`. Not sure if it requires updating or not, but would likely lead to less confusion later.

README.md
```diff
@@ -1,3 +1,226 @@
...
+and is based on [Express middleware][express], which is defined as:
+
+```
+fn(request, response, next): response
+```
+
+Based on the middleware implementations already used by frameworks that have
+adopted this signature, the following commonalities are observed:
+
+* The middleware is defined as a [callable][php-callable] using `__invoke`.
```
weierophinney: Yes, but sometimes also as a closure, function, or other instance method. Usage of `__invoke()` is not required by most of the various PSR-7 dispatchers I've seen.

README.md
```diff
@@ -1,3 +1,226 @@
...
+```
+fn(request, next): response
+```
+
+Middleware taking this approach generally has the following commonalities:
+
+* The middleware is defined as a [callable][php-callable] using a [closure][php-closure]
+  or a class with an `__invoke` method.
+* The middleware is passed 2 arguments during invocation:
+  1. A `RequestInterface` implementation.
```
weierophinney: Again, wondering if this should be `ServerRequestInterface`.

README.md
```diff
@@ -1,3 +1,226 @@
...
+[Guzzle middleware](http://docs.guzzlephp.org/en/latest/handlers-and-middleware.html)
+uses a rather unique approach to middleware that is based around generators and
+promises. Ignoring the generator portion, Guzzle middleware has the signature:
+
+```
+function(RequestInterface $request, array $options): ResponseInterface
+```
+
+[Laravel middleware](https://laravel.com/docs/master/middleware) does make use
+of HTTP Messages but supports middleware with the signature:
```
weierophinney: Are you trying to indicate it makes use of _PSR-7_ HTTP Messages (which is what you appear to mean everywhere else int the document when you use the verbiage "HTTP Messages")? I'm thinking you meant to say "does not" here...

weierophinney (changes requested):
A few minor verbiage changes, and a couple of questions.

README.md
shadowhand: Correct. This PR should be accepted as-is, it is just a copy of the existing meta document.

shadowhand:
@http-interop/http-middleware-contributors this now contains the changes from #25. Please review and let me know if you have concerns.

weierophinney:
This looks good. The only change I'd like to see is some verbiage indicating that the delegate defines the same method as middleware in part to prevent creating an object that can serve both roles. Nice work, @shadowhand !

shadowhand:
@weierophinney added.

weierophinney (approved):
:+1:

PR #27, opened 2016-11-06
Remove client middleware from scope

shadowhand:
As per the discussion [1] on the FIG mailing list, attempting to define
a specification that work for client side middleware is ill advised.

It will be far better to address client middleware in a separate
proposal after the async proposal has been completed.

[1]: https://groups.google.com/d/msg/php-fig/vBk0BRgDe2s/GTaT0yKNBgAJ


shadowhand:
cc @http-interop/http-middleware-contributors

weierophinney:
I've been nervous about this change for a while, as I've been working on http-interop compatibility with Stratigility/Expressive. This approach — removing only `MiddlewareInterface` — works perfectly for me, as I do not need to change any code currently!

:+1: from me!

weierophinney (approved):

weierophinney:
@shadowhand Any chance of a new tag that contains this change?

shadowhand:
@weierophinney I'd like to tag after #25 and #26 are reviewed.

weierophinney:
@shadowhand Fair enough! I've added reviews to each now.

PR #49, opened 2016-12-12
Remove recusive middleware stack argument

schnittstabil:
The recursion argument is flawed, see https://github.com/http-interop/http-middleware/pull/41#issuecomment-263915893 and https://github.com/http-interop/http-middleware/pull/24#issuecomment-261727171

**tl;tr** Recursion itself is not bad, only endless loops are; furthermore some weird tricks are needed to achieve that – Nobody can protect stupid people to do stupid things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,9 +285,7 @@ additional information._
 Both the middleware and delegate interface define a `process` method to prevent
 misuse of middleware as delegates.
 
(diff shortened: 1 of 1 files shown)


shadowhand (approved):

shadowhand:
Thanks!

PR #51, opened 2016-12-15
Change namespace

### LINKED ISSUE #47: Naming of DelegateInterface in light of #45

atanvarno69:
Closes #47.

Option one seems most popular:
> Rename the namespace to `Psr\Http\ServerMiddleware`, and rename the `ServerMiddlewareInterface` to `MiddlewareInterface`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "http-interop/http-middleware",
-    "description": "Common interface for HTTP middleware",
+    "description": "Common interface for HTTP server-side middleware",
     "keywords": [
         "psr",
         "psr-7",
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
     },
     "autoload": {
         "psr-4": {
-            "Interop\\Http\\Middleware\\": "src/"
+            "Interop\\Http\\ServerMiddleware\\": "src/"
         }
     },
     "extra": {
--- a/src/DelegateInterface.php
+++ b/src/DelegateInterface.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Interop\Http\Middleware;
+namespace Interop\Http\ServerMiddleware;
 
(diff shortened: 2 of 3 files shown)


schnittstabil:
Maybe the composer description have to be reconsidered too:

```json
{
     "name": "http-interop/http-middleware",
     "description": "Common interface for HTTP middleware"
}
```

atanvarno69:
@schnittstabil good spot, thank you.

Updated.

shadowhand (approved):

shadowhand:
@http-interop/http-middleware-contributors this is a breaking change and packages will need to be updated to reflect the new namespace.

Tagged in 0.4.0.

Issue #56, opened 2017-01-11
"Delegate should conflict with Middleware" argument appears unfounded


schnittstabil:
[Current version](https://github.com/http-interop/http-middleware/tree/cff4fb65317875b2bccbb6b7b22eec18b5a96f04#why-does-the-delegate-conflict-with-middleware):

> ### 5.2 Delegate Design
> #### Why does the delegate conflict with middleware?
> Both the middleware and delegate interface define a `process` method to prevent misuse of middleware as delegates.

Unfortunately, the statement above is not supported by any arguments or explanations within the `README.md`. 

More precisely,  the `README.md` lacks answers to the following questions:

1. If a middleware is used as a delegate, why must that be considered a _misuse_?
2. What kind of _misuse_? Do we have a _technical term_ for that?
3. Who will be hurt by the _misuse_?
   1. Does the _misuse_ prevent middleware dispatchers from working correctly?
   2. Does the _misuse_ prevent middlewares from working correctly?
4. Why should/must we prevent those _misuses_ by a name conflict?
   1. Why does different method names **necessarily** lead to _misusable_ classes?
   2. Why does different method names **necessarily** lead to a _misuse_?
   3. Does a name conflict safely prevent us from those _misuses_?

schnittstabil:
The mentioned sentence is the very only explanation why delegate conflict with middleware and it is not supported by any arguments within the `README.md`.

@shadowhand Using your own words at https://github.com/http-interop/http-middleware/pull/57#issuecomment-272486478:

> What may be obvious to you is not necessarily obvious to others.

Hence, why did you closed this issue without any explanation?

schnittstabil:
#### Clarification

Two issues (4. i. and 4. ii.) are similar to #55. The highlighted **necessarily** is important above. Just because one **can** do something does not **necessarily** imply he will. For example:

> _Barack Obama **can** launch atomic bombs, but that does not **necessarily** imply he will!_

@shadowhand If you believe that we have to protect developers from themselves, then we have to state that clearly. For example:

> If the middleware and delegate interfaces define methods with different names, then developers may create middlewares which can also be used as delegates. We believe this would be a misuse with negative impact on the PHP ecosystem and therefore should not be allowed.

While I disagree with that statement itself, the reasons would be more comprehensible for the reader.

schnittstabil:
[...]

PR #58, opened 2017-01-11
Remove misplaced/superfluous compatibility argument.

schnittstabil:
For clarity:
```diff
 ### 5.2 Delegate Design
 …
 #### Why isn't the delegate a `callable`?

 Using an interface type hint improves runtime safety and IDE support.

-In addition, it allows compatibility with existing middleware that already defines
-an `__invoke` method.
```

Obviously, this sentence is off-topic, it is neither about delegates nor `callable`. It may support the _Why doesn't middleware use `__invoke`?_ reasons, but if we rewrite and move it there, it would be superfluous: that section already contains the compatibility argument:

> #### Why doesn't middleware use `__invoke`?
> Doing so would conflict with existing middleware that implements the
> double-pass approach and may want to implement the middleware interface.

Therefore, I suggest to simply remove it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,9 +271,6 @@ dispatcher that uses middleware implementing `ServerMiddlewareInterface`.
 
 Using an interface type hint improves runtime safety and IDE support.
 
-In addition, it allows compatibility with existing middleware that already defines
-an `__invoke` method.
-
 _See "discussion of FrameInterface" in [relevant links](#8-relevant-links) for
 additional information._
 


shadowhand:
This is effectively the same as #55 and #56

schnittstabil:
> This is effectively the same as #55 and #56

No, it's not! The issue here is, that **the sentence is off-topic**, isn't it? – It is neither about delegates nor `callable`.

shadowhand (approved):

shadowhand:
You're right, this statement doesn't make sense in this context, since a delegate cannot be a middleware.

PR #60, opened 2017-01-14
Clarify DelegateInterface purpose

schnittstabil:
This PR reflects long discussions about the delegate naming and removes implementation details.

Related #54 and #57.

@shadowhand I truly believe this PR completely aligns with you intentions, hence please do not close it prematurely – maybe I/we can fix remaining issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,9 +263,9 @@ additional information._
 
 ### 5.2 Delegate Design
 
-The `DelegateInterface` defines a single method that accepts a request and
-returns a response. The delegate interface must be implemented by any middleware
-dispatcher that uses middleware implementing `MiddlewareInterface`.
+A delegates is a representative of the middleware dispatcher, whose only purpose
+is to encapsulate the remaining request processing. The `DelegateInterface` defines a
+`process` method therefor, which accepts a request and returns a response.
 
 #### Why isn't the delegate a `callable`?
 
@@ -278,8 +278,6 @@ additional information._
 
 Both the middleware and delegate interface define a `process` method to prevent
 misuse of middleware as delegates.
-
-The implementation of delegate should be defined within middleware dispatching systems.
 
 6. People
 ---------


schnittstabil:
Closing because of https://github.com/http-interop/http-middleware/pull/59#issuecomment-272647942

schnittstabil:
@shadowhand May you like to review this PR before I will create a new PR against fig-standards?

README.md
```diff
@@ -263,9 +263,9 @@ additional information._
 
 ### 5.2 Delegate Design
 
-The `DelegateInterface` defines a single method that accepts a request and
-returns a response. The delegate interface must be implemented by any middleware
-dispatcher that uses middleware implementing `MiddlewareInterface`.
+A delegates is a representative of the middleware dispatcher, whose only purpose
+is to encapsulate the remaining request processing. The `DelegateInterface` defines a
+`process` method therefor, which accepts a request and returns a response.
```
shadowhand, line 268: This change provides some insight, but the language is totally different than the "Middleware Design" section. Perhaps something more like:

> The `DelegateInterface ` defines a single method that accepts a server request and must return a response. This is done by dispatching the next available middleware, or if none is available, by creating an empty response.

schnittstabil: > but the language is totally different

I rarely see problems in changing writing style.

---

> This is done by dispatching the next available middleware, or if none is available, by creating an empty response.

Or throw an exception, or … – I believe such an list will always be incomplete.

Perhaps, we may just add examples, similar to the middleware section:

> _as already suggested:_
> A delegates is a representative of the middleware dispatcher, whose only purpose
> is to encapsulate the remaining request processing. The `DelegateInterface` defines a
> `process` method therefor, which accepts a request and returns a response.
> 
> _now the new part:_
> The delegate may:
> - Dispatching the next available middleware.
> - Dispatching a specific middleware, depending on the provided request.
> - Creating an empty response.
> - Creating an response, depending on the provided request.
> - …

For example:
* The second list item also covers router middleware dispatcher.
* The forth item also covers preconfigured XML/JSON responses, based on the `Accept` header of the request.

---

Btw, I would like to preserve the concrete `process` name method, it makes the following discussions easier to grasp. For example the _Why does the delegate conflict with middleware?_ section – If the reader already knows that the method name is also `process` we gain readability in my opinion.

---

While the language might be different, I like:
> whose (only) purpose is to _encapsulate the remaining request processing_

The alternative I can imagine, would be something like: 
> whose (only) purpose is to _delegate requests (back?) to the middleware dispatcher_ :disappointed:

schnittstabil:
@shadowhand Does _wrap_ sounds better? 
> whose (only) purpose is to **wrap** the remaining request processing

(_wrap_ is used at proposed/http-middleware/middleware.md in _1. Specification_)

PR #69, opened 2017-03-23
Replace readme with a link to PSR-15

shadowhand:
Trying to maintain the proposal in two different places is cumbersome.


shadowhand:
@http-interop/http-middleware-contributors anyone object to this?

PR #70, opened 2017-07-24
Replace delegate with handler interface

damianopetrungaro:


shadowhand:
I like this. Going to hold it until php-fig/fig-standards#926 is merged.

src/HandlerInterface.php
```diff
@@ -5,14 +5,14 @@
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
-interface DelegateInterface
+interface HandlerInterface
```
shadowhand, line 8: This is out of sync with the merged PSR changes. I am merging #71 instead.

PR #71, opened 2017-08-05
Update interfaces to the latest version of PSR-15

mnapoli:
Propagate changes from https://github.com/php-fig/fig-standards/pull/926 to this repository.

This is a breaking change so it needs to be 0.5.0.


shadowhand:
Possibly a duplicate of #70

mnapoli:
Yep, I'm fine if #70 is merged instead, I just want to have a PR that's ready and that will not necessitate back and forths, I've simply copy-pasted the code straight from the PSR (that explains the modified phpdoc for example).

shadowhand (approved):

shadowhand:
Released in version 0.5.0.

PR #75, opened 2017-11-06
Sync up with current PSR documentation

shadowhand:
Refs php-fig/fig-standards#950, php-fig/fig-standards#952.


weierophinney (approved):

shadowhand:
These changes have been merged into https://github.com/http-interop/http-request-handlers

